PR #27, opened 2021-04-07
Extend the README

mikhailshilkov:


sdk/python/README.md
```diff
@@ -1,3 +1,91 @@
 # Native Google Cloud Pulumi Provider (preview)
 
-The native Google Cloud Provider for Pulumi lets you use Google Cloud resources in your cloud programs.
+The native Google Cloud Provider for Pulumi lets you provision Google Cloud resources in your cloud programs.
```
viveklak, line 3: Is there a requirement to have an sdk/python/README.md? Wasn't sure why we create it generally.

README.md
```diff
@@ -1,3 +1,91 @@
...
+Credentials configuration is compatible with the classic GCP provider.
+
+Please refer to [this quickstart guide](
+https://www.pulumi.com/docs/intro/cloud-providers/gcp/setup/) for possible configuration options.
+
+## Building
+
+### Dependencies
+
+- Go 1.15
```
viveklak, line 64:
```suggestion
- Go 1.16
```
Looks like we are already on go 1.16

viveklak (approved):

sdk/python/README.md
stack72: The Python README is the one that shows in pypi - so we *generally* copy it 🤷

sdk/python/README.md
```diff
@@ -1,3 +1,91 @@
...
+Credentials configuration is compatible with the classic GCP provider.
+
+Please refer to [this quickstart guide](
+https://www.pulumi.com/docs/intro/cloud-providers/gcp/setup/) for possible configuration options.
+
+## Building
+
+### Dependencies
+
+- Go 1.15
```
viveklak, line 64:
```suggestion
- Go 1.16
```

mikhailshilkov: It's currently Go 1.15 in the SDK

PR #28, opened 2021-04-07
Add a testable Cloud Run example

mikhailshilkov:


stack72 (approved):
Going to merge so that we can get this working as part of credentials

PR #36, opened 2021-04-08
Add a Cloud Function example

mikhailshilkov:
Also, fixes the DELETE for storage bucket/object

### DIFF
--- a/examples/examples_test.go
+++ b/examples/examples_test.go
@@ -3,8 +3,14 @@
 package examples
 
 import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"net/http"
 	"os"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
 )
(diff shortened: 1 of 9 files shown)


examples/examples_test.go
```diff
@@ -29,3 +35,88 @@ func skipIfShort(t *testing.T) {
...
+				return check(bodyText)
+			}
+		}
+		if now.Sub(startTime) >= maxWait {
+			fmt.Printf("Timeout after %v. Unable to http.get %v successfully.", maxWait, hostname)
+			return false
+		}
+		count++
+		// delay 10s, 20s, then 30s and stay at 30s
+		if sleep > 30 {
```
viveklak, line 105: Maybe just replace with "github.com/jpillora/backoff" as we do in the provider?

provider/pkg/provider/provider.go
```diff
@@ -510,7 +510,7 @@ func (k *googleCloudProvider) Delete(ctx context.Context, req *rpc.DeleteRequest
 	}
 
 	id := req.GetId()
-	uri := fmt.Sprintf("%s/%s", res.BaseUrl, id)
+	uri := fmt.Sprintf("%s%s", res.BaseUrl, id)
```
viveklak, line 513: Perhaps just use this?
```suggestion
	uri := path.Join(res.BaseUrl, id)
```

viveklak (approved):

examples/examples_test.go
mikhailshilkov: I copied this 1:1 from the examples repo. We could improve both I guess.

provider/pkg/provider/provider.go
mikhailshilkov: It turns out `path.Join` doesn't quite work for URLs with all kinds of weirdness re-encoding the strings. I refactored this to a function on the resource and made some extra slashes trimmed just in case.

Issue #38, opened 2021-04-09
Support resources that are created with a PUT


mikhailshilkov:
We require the creation operation to be a POST right now, but some resources (e.g. in Pubsub) have a PUT as a create operation. We should store this in metadata and send the right HTTP request. Also, add a test and make sure that this works.

PR #39, opened 2021-04-09
Require GET methods on resources

mikhailshilkov:
In master, there are 42 resources with `NoGet: true`. I tried classifying them and found three groups:

1. Get is called "read"
2. There's no GET but there is a "list" that returns a collection
3. There's no read operation at all

None of the resources in (3) seemed obviously important, so I went ahead and changed the codegen to require a GET operation for each resource. I also added `read` as a GET operation and added a warning for skipped resources with a `list`.

On top of that, I added checks that a get operation is always an HTTP GET and create is always a POST. Added #38 to relax the latter.


provider/cmd/pulumi-resource-gcp-native/metadata.json
```diff
@@ -30666,76 +30075,6 @@
             ],
             "noDelete": true
         },
-        "gcp-native:pubsub/v1:Subscription": {
```
viveklak, line 30669: Bummer to loose some of the pubsub resources. Curious if you see anything in the spec which might allow them to be modeled?
The rest seem like a good tradeoff though!

viveklak (approved):

lblackstone (approved):

provider/cmd/pulumi-resource-gcp-native/metadata.json
mikhailshilkov: FWIW, they didn't work. It's tracked in https://github.com/pulumi/pulumi-gcp-native/issues/38

PR #41, opened 2021-04-12
Delete for non-deletable resources

mikhailshilkov:
I went through the list of resources that we mark as "NoDelete" and found only one case of a different naming: `dropDatabase` for Spanner (added an exception). There are also some "cancel", "disable", etc. methods that we may want to invoke as deletes - but they are all single cases without any apparent rule.

I found that the classic GCP provider already has 10 resources that are only deleted from the state but not from GCP - so I went ahead and added a warning of this kind to the code. We can refine this later after the initial preview.

Finally, I added a special case for IamPolicy resources: invoke an update with an empty body. IamPolicy resources have awkward IDs ending with `:getIamPolicy`. I coded around this in the provider code for now, let's see if there are any other resources like this and find an encoding in metadata if needed.

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -460,6 +460,10 @@ func (k *googleCloudProvider) Update(ctx context.Context, req *rpc.UpdateRequest
 	}
 
 	uri := res.RelativePath(req.GetId())
+	if strings.HasSuffix(uri, ":getIamPolicy") {
+		uri = strings.ReplaceAll(uri, ":getIamPolicy", ":setIamPolicy")
+	}
+
 	op, err := sendRequestWithTimeout(ctx, res.UpdateVerb, uri, body, 0)
 	if err != nil {
 		return nil, fmt.Errorf("error sending request: %s: %q %+v", err, uri, body)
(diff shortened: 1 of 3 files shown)


provider/pkg/provider/provider.go
```diff
@@ -460,6 +460,10 @@ func (k *googleCloudProvider) Update(ctx context.Context, req *rpc.UpdateRequest
 	}
 
 	uri := res.RelativePath(req.GetId())
+	if strings.HasSuffix(uri, ":getIamPolicy") {
```
viveklak, line 463: Would it make sense to set a separate `UpdatePath` field in the metadata to encode exceptions such as these where the IdPath can't be used for updates?

provider/pkg/provider/provider.go
```diff
@@ -501,12 +505,31 @@ func (k *googleCloudProvider) Delete(ctx context.Context, req *rpc.DeleteRequest
...
+			return nil, errors.Wrapf(err, "waiting for completion")
+		}
+
 		return &empty.Empty{}, nil
 	}
 
-	uri := res.RelativePath(req.GetId())
+	if res.NoDelete {
+		// At the time of writing, the classic GCP provider has the same behavior and warning for 10 resources.
+		glog.V(4).Infof("%q resources"+
```
viveklak, line 528: We seem to have a lot of NoDelete resources (270+). This warning makes sense regardless for noDelete resources, but we might be over estimating these types of resources.

mikhailshilkov: The majority of them (~200) are IamPolicies. I agree we may not actually want to generate some of those resources as resources.

provider/pkg/provider/provider.go
mikhailshilkov: I started with that but I thought it was a bit over-engineered for the single use case that we have right now. WDYT?

viveklak: Understood. Cut a ticket so we can revisit?

viveklak (approved):

provider/pkg/provider/provider.go
mikhailshilkov: https://github.com/pulumi/pulumi-gcp-native/issues/42

PR #45, opened 2021-04-15
More nuanced type spec generation for object type

viveklak:
Currently any type where the discovery doc labels the nested type in a list as `object` we emit an object type in schema which results in `map[string]string` or similar generated code. However, it seems fairly common for discovery docs to include inline types which easily break the `string->string` requirement. This change attempts to detect this and replace such types with `any` type instead.

I also included an example which demonstrates the problem. Without the change, the `allowed` block in the firewall would not compile – since the properties `IPProtocols` and `ports` expect different types.

Fixes: https://github.com/pulumi/pulumi-gcp-native/issues/18

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -32,6 +32,19 @@ func TestFunctionsTs(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestWebserverTs(t *testing.T) {
+	test := getJSBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir: filepath.Join(getCwd(t), "webserver-ts"),
+			ExtraRuntimeValidation: func(t *testing.T, stack integration.RuntimeValidationStackInfo) {
+				assertHTTPMatchesContent(t, stack.Outputs["instanceIP"].(string), "Hello, World!\n", nil)
+			},
+			SkipRefresh: true,
+		})
+
+	integration.ProgramTest(t, &test)
+}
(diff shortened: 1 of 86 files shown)


mikhailshilkov:
Thank you for finding this! Our codegen is definitely wrong here.

However, I don't think we should generate `any` in this case. Instead, we should define a type like `FirewallAllowed` for your example which would contain two properties: `IPProtocol` and `ports`. That's what we do in Azure-Native [here](https://github.com/pulumi/pulumi-azure-native/blob/d3cd290384c7c7963921113444d73e4c38432497/provider/pkg/gen/schema.go#L1318-L1319).

viveklak:
> Thank you for finding this! Our codegen is definitely wrong here.
> 
> However, I don't think we should generate `any` in this case. Instead, we should define a type like `FirewallAllowed` for your example which would contain two properties: `IPProtocol` and `ports`. That's what we do in Azure-Native [here](https://github.com/pulumi/pulumi-azure-native/blob/d3cd290384c7c7963921113444d73e4c38432497/provider/pkg/gen/schema.go#L1318-L1319).

Cool - I was wondering if we should do a type generation instead but I was wondering if there is a concern around running into possible collisions with schema types? Do we want to detect possible collisions?

mikhailshilkov:
> Do we want to detect possible collisions?

Yes, likely we do. Collisions are often tricky to debug so it's better to add a check and panic if we find them.

mikhailshilkov:
Superseded by #53

PR #53, opened 2021-04-17
Generate inline types based on object properties

mikhailshilkov:
Currently, any type where the discovery doc labels the nested type in a list as `object`, we emit an object type in the schema which results in `map[string]string` or similar generated code.

This PR respects the properties that are defined inline in the object and generates a schema type using a name like `TypeProperty`. We have a check to fail generation if such name clashes with a built-in type name.

Also, adds a simple .NET example which didn't work before this change.


provider/pkg/gen/schema.go
```diff
@@ -490,6 +490,31 @@ func (g *packageGenerator) genTypeSpec(prop *discovery.JsonSchema, isOutput bool
 		}, nil
 	case prop.Type == "any":
 		return &schema.TypeSpec{Ref: "pulumi.json#/Any"}, nil
+	case prop.Type == "object" && len(prop.Properties) > 0:
+		tok := fmt.Sprintf(`%s:%s:%s%s`, g.pkg.Name, g.mod, typeName, strings.Title(propName))
+		if isOutput {
+			tok += "Response"
+		}
+		if g.visitedTypes.Has(tok) {
```
viveklak, line 498: I think we want to check if `g.rest.schemas[fmt.Sprintf("%s%s", typeName, strings.Title(propName))]`has something instead? `visitedTypes` may not be sufficiently populated to detect a collision yet?

viveklak:
This looks great! I am happy to close https://github.com/pulumi/pulumi-google-native/pull/45 if you want to just pull in the example from there into this PR.

viveklak (approved):

provider/pkg/gen/schema.go
mikhailshilkov: Yes, great point! Fixing...

mikhailshilkov:
> pull in the example from there into this PR.

I copied the example, added random names there, and the tests are green! Thank you!

PR #60, opened 2021-04-24
docs: Make clearer it's Javascript, not Java

cobraz:
Oh, man! Thanks for this awesome provider 🎉

This pull request adds a hyphen `-` to make it clearer it's for Javascript, not Java.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ To use this package, please [install the Pulumi CLI first](https://pulumi.io/).
 
 This package is available in many languages in the standard packaging formats.
 
-### Node.js (Java/TypeScript)
+### Node.js (Java-/TypeScript)
 
 To use from JavaScript or TypeScript in Node.js, install using either `npm`:
 
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ To use this package, please [install the Pulumi CLI first](https://pulumi.io/).
 
 This package is available in many languages in the standard packaging formats.
 
-### Node.js (Java-/TypeScript)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ To use this package, please [install the Pulumi CLI first](https://pulumi.io/).
 
 This package is available in many languages in the standard packaging formats.
 
-### Node.js (Java/TypeScript)
+### Node.js (Java-/TypeScript)
```
mikhailshilkov, line 15:
```suggestion
### Node.js (JavaScript/TypeScript)
```

I guess we don't have to save keystrokes here :)

mikhailshilkov (commented):
Thank you for your attention to detail!

Could you please update [this file](https://github.com/pulumi/pulumi-google-native/blob/master/sdk/python/README.md) too? It's just a copy of the same README.

I also see that we have the same issue in almost every other Pulumi provider...

README.md
cobraz: 👍

cobraz:
> I also see that we have the same issue in almost every other Pulumi provider...

I can see if I'm able to create a simple script that adds a bunch of PRs 😅

mikhailshilkov:
> I can see if I'm able to create a simple script that adds a bunch of PRs 😅

@stack72 Do we generate READMEs in TF-based providers or are those manually edited?

mikhailshilkov (approved):

stack72:
The READMEs are manually written but they are also present in sdk/python as well in each of the providers

PR #76, opened 2021-05-07
Clean up TODOs and file issues for them

mikhailshilkov:


provider/pkg/provider/provider.go
```diff
@@ -392,22 +392,6 @@ func (p *googleCloudProvider) waitForResourceOpCompletion(baseUrl string, resp m
 	}
 }
 
-func evalParam(inputs resource.PropertyMap, param string) string {
```
mikhailshilkov, line 395: This is not needed since we moved to metadata

Issue #89, opened 2021-05-12
Incorrect args for project creation


Ezard:
`ProjectArgs` imported from `@pulumi/gcp-native/cloudresourcemanager/v3` has both `projectId` and `projectsId` as fields, with the former being optional and the latter being required

However, according to the [discovery document](https://cloudresourcemanager.googleapis.com/$discovery/rest?version=v3) and also the [API docs](https://cloud.google.com/resource-manager/reference/rest/v3/projects#Project), `projectsId` should not exist and `projectId` should be required

mikhailshilkov:
The `projectsId` property is part of the URL path `v3/projects/{projectsId}`. We will likely rename it to `projectId` and merge with the body property in the future.

> projectId should be required

Could you point me to where it's marked as required? I only see this:

```json
"projectId": {
    "description": "Immutable. The unique, user-assigned id of the project. It must be 6 to 30 lowercase ASCII letters, digits, or hyphens. It must start with a letter. Trailing hyphens are prohibited. Example: `tokyo-rain-123`",
    "type": "string"
},
```

Ezard:
@mikhailshilkov I was working off the basis that some fields are marked as "optional" in the docs (e.g. `parent` and `displayName`), whereas `projectId` is not

Also, based on the [gcloud CLI docs for creating projects](https://cloud.google.com/sdk/gcloud/reference/projects/create), a project ID is the only required data. When using the CLI, if you specify just a name and not a project ID then the CLI will auto-suggest a project ID for you; I'm unsure if this is also the case for the REST API

I'm not massively familiar with the discovery document format though, so I'm unsure how "required" parameters are specified

PR #96, opened 2021-05-19
Normalize parameter names to singular names similar to properties

mikhailshilkov:
There are two sources of SDK properties from discovery documents: URL parameters and schema type properties. URL parameters tend to be plural, e.g. `schemasId`, while property names tend to be singular, e.g. `schemaId`.

This PR standardizes on the singular form. Also, it removes the "Id" suffix from well-known parameters like `project`. This also gives an extra effect of merging a parameter and a property with the same meaning but different in pluralization (see the change in the `google.pubsub.v1.Schema` example).

Resolves https://github.com/pulumi/pulumi-google-native/issues/89


examples/simple/index.ts
```diff
@@ -6,88 +6,13 @@ import * as google from "@pulumi/google-native";
 const project = "pulumi-development";
 const region = "us-central1";
 
-const runName = "run-native";
```
viveklak, line 9: To be clear, removing this because this is covered by the cloudrun example?

provider/pkg/gen/utilities_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+var apiToSdkNames = map[string]string{
+	"projectsId":               "project",
+	"locationId":               "location",
+	"managedZonesId":           "managedZone",
+	"gameServerDeploymentsId":  "gameServerDeploymentId",
+	"schemaId":                 "schemaId",
+	"certificateAuthoritiesId": "certificateAuthorityId",
+	"sourceContents":           "sourceContents",
+	"dataId":                   "dataId",
```
viveklak, line 18: Nit - for completion - add `datumsId/datumId -> dataId` as well?

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
-(None)
+- Normalize parameter names to singular names similar to properties
```
viveklak, line 6: Do we need to callout more explicitly that this is a breaking change? Wasn't clear on what we were planning to do on this front while we are in preview.

viveklak (approved):

CHANGELOG.md
mikhailshilkov: We treated it light on azure native preview but maybe it's time to start doing so on google native

examples/simple/index.ts
mikhailshilkov: Yes, I removed duplication

Issue #99, opened 2021-05-24
Regional GKE clusters not possible since zone - a deprecated field - is required by the ClusterArgs class


brandonjbjelland:
Working within the Python SDK and trying some basic provisioning tasks against the container API. The specific issue I hit here is as the title describes: a class-required attribute, `zone`, means creation of any GKE clusters is necessarily zonal instead of regional which also means the autopilot cluster mode is currently impossible.

Proposed general solution: Looking at the [REST API doc](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters/create), `zone` (amongst others) is a deprecated field. In cases where we can reliably know an attribute is deprecated, the code generator should omit the field from the schema entirely.

## Expected behavior

I would expect to be able to omit the `zone` attribute and instead use `location` to specify the region to create a regional cluster.

## Current behavior

Currently, only zonal GKE clusters can be created via the Python (and I assume other), SDKs.

## Steps to reproduce

1. Try to instantiate `pulumi_google_native.container.v1beta1.ClusterArgs()` without specifying a `zone`. This will fail.
2. Give a zone.
3. With the other attributes satisfied, you now have a zonal cluster. Regional clusters are not possible to create if the `zone` attribute is given.

## Context (Environment)

I'm simply trying to create either a:
1. autopilot cluster - necessarily regional or
2. a regional standard GKE cluster

## Affected feature

The problem I hit is specific to GKE clusters but there's very much a generalized story here around deprecated fields. I haven't dug into if there's a dedicated field within a discovery API element that denotes deprecation status of a field but even in absence of this, the description of `zone` reads:

```python
        :param pulumi.Input[str] zone: [Output only] The name of the Google Compute Engine [zone](https://cloud.google.com/compute/docs/zones#available) in which the cluster resides. This field is deprecated, use location instead.
```

I think any fields that have some pattern that matches "This field is deprecated" or other known patterns of deprecated should likely be omitted from what the code generator pushes to the libraries.

mikhailshilkov:
Hi @brandonjbjelland thank you for reporting this.

I'm not claiming our behavior for `v1beta1` is correct - but I'm curious why you are using `v1beta1` instead of `v1`. I think this problem doesn't apply to our `v1` resource as shown in [this example](https://github.com/pulumi/examples/blob/master/google-native-ts-k8s-ruby-on-rails-postgresql/infra/cluster.ts#L10)

[...]

PR #101, opened 2021-05-26
More deprecated properties and parameters

mikhailshilkov:
Extend the pattern to catch more deprecation messages for properties and parameters

Resolve https://github.com/pulumi/pulumi-google-native/issues/99


provider/pkg/gen/schema.go
```diff
@@ -614,6 +614,23 @@ type propertyBag struct {
...
+	// we need that tricky exclusion list below to filter out the actual deprecations.
+	lowerDesc := strings.ToLower(description)
+	return strings.HasPrefix(lowerDesc, "deprecated") ||
+		strings.Contains(lowerDesc, ". deprecated") ||
+		strings.Contains(description, "(Deprecated") ||
+		strings.Contains(description, "Deprecated;") ||
+		strings.Contains(description, "Deprecated.") ||
+		strings.Contains(description, "[Deprecated]") ||
+		strings.Contains(description, "Deprecated in favor") ||
+		strings.Contains(description, "This field is deprecated")
```
viveklak, line 631: 😭

mikhailshilkov: Unfortunately, I feel like we'll have a lot of similar snippets eventually...

brandonjbjelland: Though my issue prompted this change, I want to throw some caution here that this is a pretty wide net and will likely catch more than desired. As an example, custom GCE images can be given a `DeprecationStatus`, one of which is - you guessed it - `DEPRECATED`. I would hedge on the side of catching too little rather than too much.

Thanks for the quick work here!

brandonjbjelland: Scanning through the file changes, I don't see any that look to be caught in error and I checked the concern above. Nevermind my comment! You guys are awesome 💯

viveklak: > Scanning through the file changes, I don't see any that look to be caught in error and I checked the concern above. Nevermind my comment! You guys are awesome 💯

Thanks for checking. This is definitely the kind of change that should get spidey-senses tingling...

mikhailshilkov: I'm happy to tune this rule down to a smaller or a more selective subset of rules that you folks feel would be relatively safe.

Obviously, I'd prefer Google to do this work and have a separate annotation... but there isn't one.

mikhailshilkov: Opened https://github.com/pulumi/pulumi-google-native/issues/102 to talk this through with Google folks

brandonjbjelland: One other scenario to consider: a new resource is introduced and errantly caught by this filter from the beginning. I don't have a good answer to that aside from:
1. a more concrete indicator upstream of deprecation status
2. generating a reviewable list of all the deprecated and unreviewed attributes.

PR #116, opened 2021-06-12
Fixes for storage transfer refresh

### LINKED ISSUE #107: TransferJob: Error 400: Project id was not specified.
We're utilizing `storagetransfer.v1.TransferJob` for creating a transfer job which syncs files from an AWS S3 bucket to a GCS bucket. Everything works as expected when using a `pulumi up --refresh` initially. The job gets created as expected.

Unfortunately, when making changes, for example deploying CloudRun services we receive a `Error 409: Conflict for resource '...': version '1622794891561214' was specified but current version is '1622797318138247'` which is somehow another problem, but we solved this by using `pulumi up --refresh` as a workaround in the past. Nevertheless, now, when using the `TransferJob`, this command doesn't work anymore as it fails with the following error: `google-native:storagetransfer/v1:TransferJob: Error 400: Project id was not specified.`.

This is the `TransferJob` definition.

```ts
const transferJob = new gcpNative.storagetransfer.v1.TransferJob(
  transferJobName,
  {
    transferJobsId: transferJobName,
    description: `${name}-s3-to-gcs-sync`,
    status: "ENABLED",
    schedule: {
      repeatInterval: "3600s", // every hour
      scheduleStartDate: {
        day: 1,
        month: 5,
        year: 2021,
      },
    },
    projectId: project,
    transferSpec: {
      awsS3DataSource: {
        awsAccessKey: {
          accessKeyId: config.getSecret("aws-access-key-id"),
          secretAccessKey: config.getSecret("aws-secret-access-key"),
        },
        bucketName: "my-bucket",
      },
      gcsDataSink: {
        bucketName: transferJobBucketName,
      },
      objectConditions: {
        includePrefixes: ["my/path/i-want-to-sync/"],
      },
      transferOptions: {
        deleteObjectsFromSourceAfterTransfer: false,
      },
    },
  }
);
```

At the moment, we're in a kind of deadlock situation as our workaround doesn't work anymore as well. Therefore, any hint is highly appreciated 🙂 

## Steps to reproduce

1. Create a `TransferJob`
2. Perform a `pulumi up --refresh`. Everything works as expected.
3. Perform a `pulumi up --refresh` again.
4. See error.

Expected: No `Error 400: Project id was not specified.` error

Actual: `Error 400: Project id was not specified.`

viveklak:
Fixes #107 

This change adds some additional handling to support query params in the idpath. As it turns out, TransferJob is the only resource so far to exhibit this. We could also special case this resource but this change should make us more future proof to other resources that follow this pattern.

I also restructured the wire logging in this change since the previous mechanism didn't seem to work at all and was invaluable in debugging this issue. Currently mirroring what we do in azure-native. This can serve as the stand-in till we implement a more generic wire log debugging approach as mentioned in https://github.com/pulumi/pulumi-azure-native/issues/886.


examples/storagetransfer-ts/index.ts
```diff
@@ -0,0 +1,112 @@
...
+const bucket = new storage.Bucket("my-bucket", {
+    name: bucketName,
+    project: project,
+});
+
+const storageTransferServiceAccount = gcp.storage.getTransferProjectServieAccount({
+    project: project,
+});
+
+const iam = new gcp.storage.BucketIAMMember("iam-dest", {
```
viveklak, line 31: I was trying to create this with the native provider using the following code but :
```
const iam = new storage.BucketIamPolicy("iam-dest", {
    bucket: bucket.name,
    bindings: [{
        members: [pulumi.interpolate `serviceAccount:${storageTransferServiceAccount.then(sa => sa.email)}`],
        role: "roles/storage.admin",
    }]
});
```

but this overwrites the members and makes the bucket unusable. Am I doing something wrong @mikhailshilkov? Otherwise will open an issue to track this.

viveklak:
@stack72 any chance you could enable storage transfer service on the CI project? New test failing at the moment:
```
Error when reading or editing Google Cloud Storage Transfer service account not found: googleapi: Error 403: Storage Transfer API has not been used in project 895284651812 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/storagetransfer.googleapis.com/overview?project=895284651812 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry., accessNotConfigured
```

mikhailshilkov:
@viveklak I think I enabled it now; re-running the tests

examples/storagetransfer-ts/index.ts
```diff
@@ -0,0 +1,112 @@
...
+const bucketName = pulumi.interpolate`pulumi-transfer-test-dst-${randomString.result}`;
+const transferJobName = "transfer-test";
+
+// Create a Google Cloud resource (Storage Bucket)
+const bucket = new storage.Bucket("my-bucket", {
+    name: bucketName,
+    project: project,
+});
+
+const storageTransferServiceAccount = gcp.storage.getTransferProjectServieAccount({
```
mikhailshilkov, line 27: I've just merged invokes - can you try replacing it with the native one?

P.S. `getTransferProjectServieAccount` - do we have a typo `Servie` in this GCP resource?

provider/pkg/gen/schema.go
```diff
@@ -346,12 +346,32 @@ func (g *packageGenerator) genResource(typeName string, dd discoveryDocumentReso
...
+					if err != nil {
+						return err
+					}
+				} else {
+					_, err = queryParams.WriteString("&")
+					if err != nil {
+						return err
+					}
+				}
+				queryParams.WriteString(fmt.Sprintf("%[1]s={%[1]s}", param))
```
mikhailshilkov, line 365: This feels a bit low-level. Consider applying something like https://github.com/pulumi/pulumi-google-native/blob/2da011f7ce69fcf7e2c05f2f2ce0777617b31f2a/provider/pkg/provider/http.go#L227 instead.

provider/pkg/gen/schema.go
```diff
@@ -346,12 +346,32 @@ func (g *packageGenerator) genResource(typeName string, dd discoveryDocumentReso
...
+			}
+
+			params := queryParams.String()
+			v, err := g.buildIdParams(typeName, idPath, params, inputProperties, &response)
```
mikhailshilkov, line 369: It feels like it would be cleaner to append `params` to `idPath` and pass it as a single URL to `buildIdParams`. Would it be possible to amend `buildIdParams` to detect the name as the last parameter in the path, excluding query? Or change it to be the last unknown parameter (`projectId` would be known).

Just something to consider.

examples/examples_nodejs_test.go
```diff
@@ -55,6 +55,18 @@ func TestWebserverTs(t *testing.T) {
...
+func TestStorageTransferTs(t *testing.T) {
+	test := getJSBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir:         filepath.Join(getCwd(t), "storagetransfer-ts"),
+			SkipRefresh: true,
+		})
+
+	integration.ProgramTest(t, &test)
+	test.SkipRefresh = false
+	integration.ProgramTest(t, &test)
```
mikhailshilkov, line 67: Consider adding a comment about why we do it like this.

mikhailshilkov (commented):
Thank you for finding this bug and for adding logging!

mikhailshilkov:
I think it's a different build error now

examples/storagetransfer-ts/index.ts
viveklak: Cut https://github.com/pulumi/pulumi-google-native/issues/118

examples/storagetransfer-ts/index.ts
viveklak: Yup - https://github.com/pulumi/pulumi-gcp/blob/master/provider/resources.go#L2333

examples/storagetransfer-ts/index.ts
```diff
@@ -0,0 +1,112 @@
+import * as gcp from "@pulumi/gcp";
```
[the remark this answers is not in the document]
viveklak: Fixed.

examples/storagetransfer-ts/index.ts
viveklak: I tried to use the new invokes but as highlighted in #119, we are not exposing the project's service-specific service accounts getters. I tried to use the generic IAM getServiceAccount but it doesn't seem to be able to enumerate the transfer service account. I verified this from the IAM console as well (although I see some service-specific service accounts for the project for other services). For now, I am leaving this in as is.

provider/pkg/gen/schema.go
viveklak: I restructured it a bit. LMK what you think.

examples/examples_nodejs_test.go
viveklak: I was doing this wrong. I fixed it now. I forgot we do an empty refresh anyway by default.

provider/pkg/gen/schema.go
viveklak: See the latest update which should reflect the former suggestion you made. Let me know what you think.

viveklak:
> I think it's a different build error now. Also, please rebase on master to avoid those extensive `schema-tools` warnings.

Fixed now.

mikhailshilkov (approved):
This looks great, thank you for finding and fixing the issue!

PR #140, opened 2021-07-01
Update pulumi dependencies to v3.6.0

justinvp:


sdk/go/google/accesscontextmanager/v1/accessLevel.go
```diff
@@ -59,29 +59,9 @@ func GetAccessLevel(ctx *pulumi.Context,
 
 // Input properties used for looking up and filtering AccessLevel resources.
 type accessLevelState struct {
-	// A `BasicLevel` composed of `Conditions`.
```
justinvp, line 62: @pgavlin, these changes to state are expected, right?

sdk/python/pulumi_google_native/apigateway/v1/api.py
```diff
@@ -17,12 +17,12 @@ def __init__(__self__, *,
                  location: pulumi.Input[str],
                  project: pulumi.Input[str],
                  display_name: Optional[pulumi.Input[str]] = None,
-                 labels: Optional[pulumi.Input[Mapping[str, pulumi.Input[str]]]] = None,
+                 labels: Optional[pulumi.Input[Mapping[str, str]]] = None,
```
justinvp, line 20: This change looks suspicious. I haven't tracked down why. @mikhailshilkov, is this an expected change? If not, it may be a regression from https://github.com/pulumi/pulumi/pull/7059 cc: @pgavlin

justinvp: Looks like there are similar changes to the Node.js SDK.

mikhailshilkov: I wouldn't expect this, no. Did it not happen in other providers?

sdk/go/google/accesscontextmanager/v1/accessLevel.go
pgavlin: Yes.

sdk/python/pulumi_google_native/apigateway/v1/api.py
pgavlin: This is definitely not expected. What does the schema look like for this property?

mikhailshilkov: It's defined as

```
"labels": {
  "type": "object",
  "description": "Optional. Resource labels to represent user-provided metadata. Refer to cloud documentation on labels for more details. https://cloud.google.com/compute/docs/labeling-resources"
},
```

mikhailshilkov: Are we tracking this issue somewhere?

pgavlin: We are not. I just tried a standalone repro for this locally and got the output I would expect: `Optional[pulumi.Input[Mapping[str, pulumi.Input[str]]]]`

I'll try pulling this branch down and reproing within context.

pgavlin: Okay, I can repro this locally. Digging in now.

pgavlin: Alright, I have a fix.

pgavlin: Fix is part of https://github.com/pulumi/pulumi/pull/7475

mikhailshilkov:
Closing in favor of https://github.com/pulumi/pulumi-google-native/pull/152

PR #156, opened 2021-07-17
Create new client with refreshed credentials

### LINKED ISSUE #138: Ensure that renewed credentias are detected by the provider
Google native provider can be configured to use static service account credentials or short-lived auth tokens (typically expire in 1 hr). Currently the provider configures a static client for the lifetime of the provider at initial configuration time. While this works for static credentials, for long running provisioning sessions with short-lived tokens this becomes problematic. Some of the more advanced auth mechanisms such as using [identity federation with an OIDC provider](https://cloud.google.com/iam/docs/access-resources-oidc#generate-automatic) rely on settings such as APPLICATION_DEFAULT_CREDENTIALS being updated out-of-band which runs counter to a statically configured client for the lifetime of the provider.

Instead of configuring a [statically configured client ](https://github.com/pulumi/pulumi-google-native/blob/a3085c1fc87978fbb5967fcee2f2650fbf9bf1ef/provider/pkg/provider/provider.go#L125) we could use a simple factory to intercept access to the client which can:
a. vend an existing client if its oauth token is not expired
b. create a fresh client if the existing client’s oauth token is expired and update the factory instance with this

When creating a new client in `b.` the credential discovery can be revisited to pull in any changes - e.g. through the metadata server or an external process updating credential files etc. 
**Note** - this will only work for file-backed credentials. Tokens or credentials passed explicitly as environment variables can't be changed for the lifetime of the process. #153 might allow more flexibility here.

viveklak:
Fixes #138 

This change updates the client used by the native provider to be aware of the validity of its underlying tokens/credentials. If invalid, the credentials are refreshed from the various sources. e.g. files containing credentials or tokens are reloaded, environment variables re-evaluated etc. This allows the provider to play a bit better with short lived credentials as specified in #138.


provider/pkg/googleclient/credentials.go
```diff
@@ -0,0 +1,171 @@
...
+	defaultTS, err := google.DefaultTokenSource(context.Background(), c.Scopes...)
+	if err != nil {
+		return nil, fmt.Errorf("Attempted to load application default credentials since neither `credentials` nor `access_token` was set in the provider block.  No credentials loaded. To use your gcloud credentials, run 'gcloud auth application-default login'.  Original error: %w", err)
+	}
+	return &google.Credentials{
+		TokenSource: defaultTS,
+	}, err
+}
+
+// TODO: Why do we need this?
```
mikhailshilkov, line 167: Isn't it used in `getCredentials` above? Or what is the question?

provider/pkg/googleclient/credentials.go
```diff
@@ -0,0 +1,171 @@
...
+	return g(ctx, c)
+}
+
+func defaultCredentialRetriever() credentialRetriever {
+	return getCredentialsFunc(getCredentials)
+}
+
+func getCredentials(ctx context.Context, c Config) (*google.Credentials, error) {
+	if c.AccessToken != "" {
+		contents, _, err := pathOrContents(c.AccessToken)
```
mikhailshilkov, line 86: Somehow I missed that we already support reading from the file!

mikhailshilkov (approved):
This makes total sense! I didn't realize we were that close to a working solution, thanks for digging it!

provider/pkg/googleclient/credentials.go
viveklak: > Isn't it used in `getCredentials` above? Or what is the question?

My specific question was around this comment: `staticTokenSource is used to be able to identify static token sources without reflection.`
I didn't see any introspection on the token source - wonder if this a hold over from the TF provider?

mikhailshilkov: Yes, probably

viveklak: I removed this type for now since we don't have great use for it.

Issue #124, opened 2021-06-16
Custom mutation support for container resources


viveklak:
We should cover more update scenarios and make sure the replacement of clusters/nodepools is only triggered when necessary (i.e. the immutable properties of these resources are accurately tracked).

mikhailshilkov:
This likely needs to be a manual implementation of Update methods (and corresponding Diff yielding no replacements). 

Cluster is relatively simple: we'd need to put updated properties inside an `update` map and prepend `desired` to property names.

Node pool is quite complicated as it has separate methods to update several groups of properties. See [this function](https://github.com/hashicorp/terraform-provider-google-beta/blob/889a6c013e7757b0f0592fe4ca275a781db9af5c/google-beta/resource_container_node_pool.go#L809) in the TF provider to get a sense of the complexity.

I suggest declaring some custom "updaters" that would apply to some resources, would declare which properties it can mutate, and then produce the payloads and actual calls with HTTP or Go SDK.

Separate attention should be given to awaiting of ready states, as the TF provider does quite a few of those.

Any better ideas @viveklak ?

MitchellGerdisch:
Presumably this issue would also support loop-based creation of nodegroups.
Currently in GCP classic, I can have code like this:
```
for idx in range(2):
    np_name = f"basename-{idx}"
    node_pool = gcp.container.NodePool(np_name,
         cluster=gke_cluster.cluster.name,
         name=np_name,
         project=project,
         location=region,
     )
```
Because the classic provider nodepool creation logic sequences the nodepool creations.
But when trying this with google-native, one gets an 400 back from GCP about the second nodepool indicating that a nodepool creation is already in progress and to try again later.

viveklak:
This issue will unblock https://github.com/pulumi/pulumi-google-native/issues/263 as well.

PR #169, opened 2021-07-27
Support updates for GKE clusters

viveklak:
Partial fix for #124. This only handles cluster update and is still missing some additional fields that can be updated on the cluster. See comments in the PR for the current limitations.


provider/pkg/provider/handlers.go
```diff
@@ -0,0 +1,98 @@
...
+	case "google-native:container/v1:Cluster", "google-native:container/v1beta1:Cluster":
+		return updateHandlerFunc(func(inputs, oldState resource.PropertyMap) (map[string]interface{}, error) {
+			res, ok := p.resourceMap.Resources[resourceTok]
+			if !ok {
+				return nil, errors.Errorf("resource %q not found", resourceTok)
+			}
+
+			updateBody := p.prepareAPIInputs(inputs, oldState, res.UpdateProperties)
+
+			// TODO: Cluster updates also include flattened members of the `networkConfig` map
```
viveklak, line 39: See https://github.com/pulumi/pulumi-google-native/blob/master/discovery/container_v1.json#L3029 for examples of fields that are originally specified under `networkConfig` in create properties but are flattened in the update request.

In addition, the update also is not sufficient to adequately cover all mutable fields. See `updateMaster` for an example: https://github.com/pulumi/pulumi-google-native/blob/master/discovery/container_v1.json#L3029

provider/pkg/provider/handlers.go
```diff
@@ -0,0 +1,98 @@
...
+					if strings.HasPrefix(k, "desired") {
+						splitBody = append(splitBody, map[string]interface{}{k: v})
+					}
+				}
+			}
+
+			logging.V(9).Infof("updateBody: %+v", updateBody)
+			logging.V(9).Infof("splitBody: %+v", splitBody)
+
+			// Update operation on GKE clusters are to be performed one field at a time.
```
viveklak, line 57: The terraform provider performs some [cluster level locking](https://github.com/hashicorp/terraform-provider-google-beta/blob/master/google-beta/resource_container_cluster.go#L1861) to prevent concurrent updates to the same cluster. Is this something we need to guard against as well?

provider/cmd/pulumi-resource-google-native/schema.json
```diff
@@ -70556,6 +70556,53 @@
                 "resourceLimits"
             ]
         },
+        "google-native:container/v1:ClusterDesiredDatapathProvider": {
```
viveklak, line 70559: Will look to avoid rename these types before finalizing PR.

viveklak:
@mikhailshilkov would appreciate an initial look at this.

provider/pkg/gen/utilities.go
```diff
@@ -42,6 +42,11 @@ func apiPropNameToSdkName(name string) string {
 	}
 }
 
+func stripDesiredPrefix(name string) string {
```
mikhailshilkov, line 45: Nit: add a comment with examples of an input and an output

provider/pkg/provider/handlers.go
```diff
@@ -0,0 +1,98 @@
...
+type updateHandlerFunc func(inputs resource.PropertyMap, oldState resource.PropertyMap) (map[string]interface{}, error)
+
+func (u updateHandlerFunc) Update(inputs resource.PropertyMap, oldState resource.PropertyMap) (map[string]interface{}, error) {
+	return u(inputs, oldState)
+}
+
+func (p *googleCloudProvider) updateHandler(resourceTok string, req *rpc.UpdateRequest) updateHandler {
+	logging.V(9).Infof("UpdateHandler for %q", resourceTok)
+	switch resourceTok {
+	case "google-native:container/v1:Cluster", "google-native:container/v1beta1:Cluster":
```
mikhailshilkov, line 30: Consider an approach similar to Azure Native does: add an extra property to resource metadata to define the update style, set it while generating metadata, and rely on it here rather than resource names. It reduces duplication of resource names and makes this an explicit property of a resource that can be tracked over time.

provider/pkg/provider/handlers.go
mikhailshilkov: What would be a scenario when we'd get a concurrent update?

provider/pkg/provider/handlers.go
```diff
@@ -0,0 +1,98 @@
...
+			// which is hard to represent with the current metadata format in place (sdkName
+			// can only be used to refer to fields at the top level).
+
+			var splitBody []map[string]interface{}
+			update, hasUpdate := updateBody["update"]
+			updateMap, isMap := update.(map[string]interface{})
+			if hasUpdate && isMap {
+				for k, v := range updateMap {
+					if strings.HasPrefix(k, "desired") {
+						splitBody = append(splitBody, map[string]interface{}{k: v})
```
mikhailshilkov, line 49: I don't quite get how `splitBody` works... Is it an array of maps, each map having exactly one key? Are we sending exactly one key in the update body? Will it keep other fields intact?

This seems very different from a "normal" update where we'd send the full payload of the updated resource, so it's worth documenting the unique behavior here.

provider/pkg/provider/handlers.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (u updateHandlerFunc) Update(inputs resource.PropertyMap, oldState resource.PropertyMap) (map[string]interface{}, error) {
+	return u(inputs, oldState)
+}
+
+func (p *googleCloudProvider) updateHandler(resourceTok string, req *rpc.UpdateRequest) updateHandler {
+	logging.V(9).Infof("UpdateHandler for %q", resourceTok)
+	switch resourceTok {
+	case "google-native:container/v1:Cluster", "google-native:container/v1beta1:Cluster":
+		return updateHandlerFunc(func(inputs, oldState resource.PropertyMap) (map[string]interface{}, error) {
```
mikhailshilkov, line 31: Would it be worth refactoring this to a stand-alone named function?

provider/pkg/gen/schema.go
```diff
@@ -519,6 +524,30 @@ func (g *packageGenerator) genResource(typeName string, dd discoveryDocumentReso
...
+			return err
+		}
+		for name, value := range updateBag.properties {
+			if _, has := resourceMeta.CreateProperties[stripDesiredPrefix(name)]; has {
+				value.SdkName = stripDesiredPrefix(name)
+				resourceMeta.UpdateProperties[name] = value
+			} else if name == "desiredClusterAutoscaling" {
+				value.SdkName = "autoscaling"
+				resourceMeta.UpdateProperties[name] = value
+			}
```
mikhailshilkov, line 544: Which properties would fall into a final `else` statement here? Should we do anything about them?

mikhailshilkov (commented):
Looks good overall. I suspect we won't be able to take a similar metadata-driven approach for Node Pools and some Compute resources, so I wonder if it makes sense for this single use case. If we have to hard-code everything for those use cases, maybe even with the Google Go SDK, should we be doing the same here?

viveklak:
Closed in favor of #588

PR #219, opened 2021-10-29
Set additional keywords in the Pulumi schema spec

praneetloke:
Add category and kind tags to the Keywords array. I would also like to update the `pulumi/pulumi/pkg` dep version and set the new  `DisplayName` and `Publisher` properties as well but my local setup seems to have all sorts of problems while trying to open this project (aws-native, azure-native too) so I'd love your help to do that separately.


mikhailshilkov (approved):
This LGTM - what kind of problems did you get locally?

praneetloke:
Closing as Levi is going to have a new PR to replace this.

PR #238, opened 2021-11-23
Disable response body dump

### LINKED ISSUE #234: .NET test failing in master
error: error sending request: http2: response body closed: "https://storage.googleapis.com/storage/v1/b?project=pulumi-ci-gcp-provider" map[name:my-bucket-3056707 project:pulumi-ci-gcp-provider]

https://github.com/pulumi/pulumi-google-native/runs/4253969472?check_suite_focus=true#step:21:144

mikhailshilkov:
Fix #234 (well, rather work around it with a temporary measure)

It turns out that the build failure was caused by CI updating to Go 1.16.10 (from 1.16.9) which shipped a breaking bug https://github.com/golang/go/issues/49366 That bug makes `httputil.DumpResponse` fail for HTTP/2 responses.

In the spirit of the simplest possible fix of P1, I've disabled dumping the response for now. It looks like they want to backport the fix to 1.16.11 - so we can probably wait for that one to release. Tracked in https://github.com/golang/go/issues/49558

### DIFF
--- a/examples/gke-ts/index.ts
+++ b/examples/gke-ts/index.ts
@@ -3,7 +3,7 @@
 import * as google from "@pulumi/google-native";
 
 // TODO: Determine this dynamically once https://github.com/pulumi/pulumi-google-native/issues/166 is done.
-const engineVersion = "1.20.9-gke.1001";
+const engineVersion = "1.20.10-gke.1600";
 
 const nodeConfig: google.types.input.container.v1.NodeConfigArgs = {
     machineType: "n1-standard-2",
--- a/provider/pkg/googleclient/http.go
+++ b/provider/pkg/googleclient/http.go
@@ -111,7 +111,9 @@ func (c *GoogleClient) RequestWithTimeout(method, rawurl string, body map[string
 		return nil, err
 	}
 
-	debugResp, _ := httputil.DumpResponse(res, true)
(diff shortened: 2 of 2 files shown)


examples/gke-ts/index.ts
```diff
@@ -3,7 +3,7 @@
 import * as google from "@pulumi/google-native";
 
 // TODO: Determine this dynamically once https://github.com/pulumi/pulumi-google-native/issues/166 is done.
-const engineVersion = "1.20.9-gke.1001";
+const engineVersion = "1.20.10-gke.1600";
```
mikhailshilkov, line 6: Unrelated - just bumping to a more modern version

viveklak:
> Fix #234 (well, rather work around it with a temporary measure)
> 
> It turns out that the build failure was caused by CI updating to Go 1.16.10 (from 1.16.9) which shipped a breaking bug [golang/go#49366](https://github.com/golang/go/issues/49366) That bug makes `httputil.DumpResponse` fail for HTTP/2 responses.
> 
> In the spirit of the simplest possible fix of P1, I've disabled dumping the response for now. It looks like they want to backport the fix to 1.16.11 - so we can probably wait for that one to release. Tracked in [golang/go#49558](https://github.com/golang/go/issues/49558)

Yikes! Thanks for digging through the mystery!

viveklak (approved):

PR #253, opened 2021-12-07
Fix link to guide

### LINKED ISSUE #250: README points to the classic provider's config
> Please refer to this quickstart guide for possible configuration options.

Points to the GCP provider which uses the `gcp:*` configuration scheme while we should point to `google-native:*`. Also, 

> Credentials configuration is compatible with the classic GCP provider.

isn't exactly right - we should point to the difference above.

danielrbradley:
Closes #250 

- Point to native guide rather than classic.
- Tidy up code blocks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,10 +43,11 @@ generate_nodejs::
 	$(WORKING_DIR)/bin/$(CODEGEN) nodejs ${VERSION}
 
 build_nodejs:: VERSION := $(shell pulumictl get version --language javascript)
+build_nodejs:: TSC := $(shell which tsc)
 build_nodejs::
 	cd ${PACKDIR}/nodejs/ && \
 		yarn install && \
-		node --max-old-space-size=4096 /usr/local/bin/tsc --diagnostics && \
+		node --max-old-space-size=4096 $(TSC) --diagnostics && \
 		cp ../../README.md ../../LICENSE package.json yarn.lock ./bin/ && \
 		sed -i.bak -e "s/\$${VERSION}/$(VERSION)/g" ./bin/package.json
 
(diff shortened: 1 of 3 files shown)


mikhailshilkov (approved):
This is great. FWIW, #250 has a second part to clarify that "Credentials configuration is compatible with the classic GCP provider." isn't exactly true: the keys should start with `google-native` instead of `gcp`. I suspect we have this problem in all `-native` providers.

danielrbradley:
> "Credentials configuration is compatible with the classic GCP provider." isn't exactly true: the keys should start with `google-native` instead of `gcp`.

Just been looking at this and I think it's slightly misleading but technically true - the **credentials** are configured in exactly the same way as the classic providers - fetched from the associated CLI tool or environment variable. But the non-credentials configuration is different as it uses Pulumi's config which is then namespaced for the provider.

danielrbradley:
@mikhailshilkov looking at the classic provider's readme - there's no information on this at all. You happy if I just remove the line completely to remove confusion?

mikhailshilkov:
> Just been looking at this and I think it's slightly misleading but technically true - the credentials are configured in exactly the same way as the classic providers 

Ah, that's fair.

> But the non-credentials configuration is different as it uses Pulumi's config which is then namespaced for the provider.

Maybe we could add a sentence about this.

> Looking at the classic provider's readme - there's no information on this at all. You happy if I just remove the line completely to remove confusion?

I guess I'd rather keep it and clarify the statement. Using two providers at the same time is an important scenario that we want to highlight.

mikhailshilkov (approved):
LGTM

PR #271, opened 2021-12-22
Serve uncompressed schema from GetSchema RPC

viveklak:


viveklak:
@mikhailshilkov we might need the same thing for the other native providers. Will make the change once this is approved.

mikhailshilkov (approved):
LGTM. I didn't realize this has been used outside converters.

viveklak:
> LGTM. I didn't realize this has been used outside converters.

Yeah - bit of an oversight given its part of our RPC surface area. I will make the changes in the other native providers.

PR #284, opened 2022-01-07
Fix naming of projectId property on the Project resource

### LINKED ISSUE #277: GetProjectResult.projectId is missing
## Hello!

- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

The `GetProjectResult` V3 interface expects `project` to be the key of the project id in the response of `v3.getProject` but the GCP server returns `projectId`. GCP's docs list the response body to contain `projectId`: https://cloud.google.com/resource-manager/reference/rest/v3/projects#Project

### Steps to reproduce

1. Set up google native typescript project.
1. Fetch existing project via `@pulumi/google-native/cloudresourcemanager/v3.getProject`
1. Run `util.inspect(response)` to see actual response. Notice `projectId` and not `project`.

Expected:  The project id be present in `response.project`
Actual:  `undefined`.
Workaround: `(response as any).projectId`

danielrbradley:
Fixes #277 

We replace projectId with project elsewhere because we then build the full projectId dynamically within the provider. However the projectId within the actual Project resource type does not need auto-naming.

Not  committing the test example as it seems to take some time to actually delete the project which could cause issues in CI/CD.

### DIFF
--- a/provider/pkg/gen/utilities_test.go
+++ b/provider/pkg/gen/utilities_test.go
@@ -40,7 +40,7 @@ var apiPropToSdkNames = map[string]string{
 
 func TestApiPropNameToSdkName(t *testing.T) {
 	for name, expected := range apiPropToSdkNames {
-		actual := apiPropNameToSdkName(name)
+		actual := apiPropNameToSdkName("typeName", name)
 		assert.Equal(t, expected, actual)
 	}
 }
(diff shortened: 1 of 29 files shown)


provider/pkg/gen/utilities_test.go
```diff
@@ -40,7 +40,7 @@ var apiPropToSdkNames = map[string]string{
 
 func TestApiPropNameToSdkName(t *testing.T) {
 	for name, expected := range apiPropToSdkNames {
-		actual := apiPropNameToSdkName(name)
+		actual := apiPropNameToSdkName("typeName", name)
```
viveklak, line 43: Nit - maybe add the `Project` type name to the tests while you are at it?

viveklak (approved):

provider/pkg/gen/utilities_test.go
danielrbradley: Have just added the exception case to the test - all other types will be treated the same.

PR #295, opened 2022-01-13
Avoid duplicate go module names

danielrbradley:
gopls complains at duplicate modules within the same workspace so make fake module names unique. This should make the project able to be opened in VSCode and other editors with no custom settings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,9 +83,9 @@ build_go::
 	cd sdk/ && go build github.com/pulumi/pulumi-google-native/sdk/go/google/...
 
 clean::
-	rm -rf sdk/nodejs && mkdir sdk/nodejs && touch sdk/nodejs/go.mod
-	rm -rf sdk/python && mkdir sdk/python && touch sdk/python/go.mod && cp README.md sdk/python
-	rm -rf sdk/dotnet && mkdir sdk/dotnet && touch sdk/dotnet/go.mod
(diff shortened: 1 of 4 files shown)


danielrbradley:
Alternatively, is there a way to just remove these files? They seem to just be added as a hack for GoLand possibly?

viveklak:
> Alternatively, is there a way to just remove these files? They seem to just be added as a hack for GoLand possibly?

I think you need the empty go.mods to prevent the go sdk including all the non-go sdks. See https://github.com/pulumi/pulumi-aws/pull/1631 for instance.

stack72 (changes requested):
These are needed to ensure that these folders are not added to the resulting go SDK I'm afraid :/ This is a go mod-ism around this

danielrbradley:
The issue I'm seeing is that the empty file causes gopls to fail - but adding a module name fixes it. Are you saying it _has_ to be an empty file? AWS Native was set up with just stub modules like I've proposed rather than empty files. Each module had a name and version so it validates as a `go.mod` file.

danielrbradley:
Possible alternative approach for fixing the core issue: avoid nesting the other sdks inside the go sdk folder.

Currently, the go SDK module is defined in `sdk/go.mod` and therefore we have to exclude `sdk/dotnet`, `sdk/nodejs` etc.

If we moved `sdk/go.mod` to `sdk/go/go.mod` then maintained paths by moving `sdk/go` into `sdk/go/go` then we'd only have to fix the paths used in the codegen and make file. The other sdks would then not sit within a go module at all.

Do you foresee any issues with this approach?

stack72 (approved):

PR #297, opened 2022-01-14
Use TypeScript package from Yarn, not global

AaronFriel:
The `tsc` command on line 52 would always use a globally installed TypeScript, regardless of the one installed by yarn on the previous line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,11 +46,10 @@ generate_nodejs::
 	$(WORKING_DIR)/bin/$(CODEGEN) nodejs ${VERSION}
 
 build_nodejs:: VERSION := $(shell pulumictl get version --language javascript)
-build_nodejs:: TSC := $(shell which tsc)
 build_nodejs::
 	cd ${PACKDIR}/nodejs/ && \
 		yarn install && \
-		node --max-old-space-size=4096 $(TSC) --diagnostics && \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -46,11 +46,10 @@ generate_nodejs::
 	$(WORKING_DIR)/bin/$(CODEGEN) nodejs ${VERSION}
 
 build_nodejs:: VERSION := $(shell pulumictl get version --language javascript)
-build_nodejs:: TSC := $(shell which tsc)
 build_nodejs::
 	cd ${PACKDIR}/nodejs/ && \
 		yarn install && \
-		node --max-old-space-size=4096 $(TSC) --diagnostics && \
+		node --max-old-space-size=4096 ./node_modules/.bin/tsc --diagnostics && \
```
t0yv0, line 52: Also possible to spell `yarn run tsc` if this ever had to run on Windows.

t0yv0:
This LGTM but I don't work regularly with this repo and its build so I think this needs a maintainer to make sure this wasn't an intentional thing to use the global TSC.

Makefile
AaronFriel: I'll look at refactoring this so we can run it consistently - the node args make this a bit trickier to do as either a run command or package.json script if I recall correctly. There is an incantation that'll do it though.

AaronFriel:
Replaced with #302 to run tests.

PR #414, opened 2022-03-25
Automerge weekly sdk update

jkisk:
Part of [#1915](https://github.com/pulumi/home/issues/1915).

### DIFF
--- a/.github/workflows/weekly-pulumi-update.yml
+++ b/.github/workflows/weekly-pulumi-update.yml
@@ -1,7 +1,7 @@
 name: weekly-pulumi-update
 "on":
   schedule:
-    - cron: 35 12 * * 4
+    - cron: "35 12 * * 4"
   workflow_dispatch: {}
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly-pulumi-update.yml
```diff
@@ -1,7 +1,7 @@
 name: weekly-pulumi-update
 "on":
   schedule:
-    - cron: 35 12 * * 4
+    - cron: "35 12 * * 4"
```
stack72, line 4: This should not be needed...

.github/workflows/weekly-pulumi-update.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         goversion: [1.17.x]
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
stack72, line 19: Is there any major changes here we need to test first before upgrading to v3?

stack72 (commented):
Few questions

.github/workflows/weekly-pulumi-update.yml
jkisk: Not that I can see, but would appreciate your thoughts: https://github.com/actions/checkout/releases/tag/v3.0.0. I bumped it so we could use the latest of https://github.com/peter-evans/create-pull-request.

PR #430, opened 2022-04-05
Automerge nightly SDK PR

jkisk:
Set PR to automerge if all checks are green, and also fix an input in the weekly pulumi/pulumi update action.  @stack72 do we need to add an additional check for breaking changes before we make this change? 
Resolves #1915


viveklak:
hmm why is the no-changelog-required label not dropping the Update Changelog check?

jkisk:
> hmm why is the no-changelog-required label not dropping the Update Changelog check?

I thought that was odd too, perhaps our org wide bot had a blip?  I removed and readded label just now and it's working.

stack72 (approved):

PR #432, opened 2022-04-07
Update author input for automated PRs

jkisk:
Add input so pulumi-bot will be the author of automated PRs

### DIFF
--- a/.github/workflows/nightly-sdk-generation.yml
+++ b/.github/workflows/nightly-sdk-generation.yml
@@ -65,17 +65,18 @@ jobs:
           git commit -m "Discovery documents"
           git add .
           git commit -m "Regenerating SDKs based on updated modules"
-          git push origin generate-sdk/${{ github.run_id }}-${{ github.run_number }}
+          git push origin generate-sdk/${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}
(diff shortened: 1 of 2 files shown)


jkodroff (approved):

viveklak:
If we rerun, don’t we want the change to be cumulative - I.e keep the same branch? Just trying to understand the problem this causes.

jkisk:
> If we rerun, don’t we want the change to be cumulative - I.e keep the same branch? Just trying to understand the problem this causes.

That's a good point actually, and I'm open to keeping the same branch name if that's better.  I'm not familiar enough with the process here to know if we would ever want to have access to the generated sdks from an unsuccessful run.

viveklak:
> > If we rerun, don’t we want the change to be cumulative - I.e keep the same branch? Just trying to understand the problem this causes.
> 
> That's a good point actually, and I'm open to keeping the same branch name if that's better. I'm not familiar enough with the process here to know if we would ever want to have access to the generated sdks from an unsuccessful run.

I suppose if we needed to we could always replicate that locally? I think reducing the branch splay is a better option. I'd recommend removing the run_attempt for now but keep the author part?

viveklak (approved):
Change the PR title as well?

Issue #658, opened 2022-08-29
Cloud DNS v2 Doesn't Work


rawkode:
### What happened?

```
  google-native:dns/v2:ManagedZone (example.com):
    error: error sending request: googleapi: got HTTP response code 404 with body: <!DOCTYPE html>
    <html lang=en>
      <meta charset=utf-8>
      <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
      <title>Error 404 (Not Found)!!1</title>
      <style>
        *{margin:0;padding:0}html,code{font:15px/22px arial,sans-serif}html{background:#fff;color:#222;padding:15px}body{margin:7% auto 0;max-width:390px;min-height:180px;padding:30px 0 15px}* > body{background:url(//www.google.com/images/errors/robot.png) 100% 5px no-repeat;padding-right:205px}p{margin:11px 0 22px;overflow:hidden}ins{color:#777;text-decoration:none}a img{border:0}@media screen and (max-width:772px){body{background:none;margin-top:0;max-width:none;padding-right:0}}#logo{background:url(//www.google.com/images/branding/googlelogo/1x/googlelogo_color_150x54dp.png) no-repeat;margin-left:-5px}@media only screen and (min-resolution:192dpi){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat 0% 0%/100% 100%;-moz-border-image:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) 0}}@media only screen and (-webkit-min-device-pixel-ratio:2){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat;-webkit-background-size:100% 100%}}#logo{display:inline-block;height:54px;width:150px}
      </style>
      <a href=//www.google.com/><span id=logo aria-label=Google></span></a>
      <p><b>404.</b> <ins>That’s an error.</ins>
      <p>The requested URL <code>/dns/v2/projects/example/locations/europe-west2/managedZones</code> was not found on this server.  <ins>That’s all we know.</ins>
    : "https://dns.googleapis.com/dns/v2/projects/example/locations/europe-west2/managedZones" map[__autonamed:true description:Managed DNS by Pulumi dnsName:example.com. location:europe-west2 name:example-f74c225 project:example]
```

### Steps to reproduce

```ts
[...]

PR #662, opened 2022-08-29
Remove DNS v2 API which is not actively advertised by Google

viveklak:
Fixes https://github.com/pulumi/pulumi-google-native/issues/658

As mentioned here: https://github.com/pulumi/pulumi-google-native/issues/658#issuecomment-1230618151 the api spec is no longer vended by discovery api, nor is it listed in the docs repo. There are some unexplained behavioral changes in DNS v2 as discovered in the above issue. It might be best to exclude this api for now.


viveklak:
Fixes #658

lblackstone (approved):

Issue #865, opened 2023-04-25
eventarc.Trigger should have no validateOnly parameter


mikhailshilkov:
### What happened?

Currently, `validateOnly` is a property of the `eventarc.Trigger` resource (and some others). However, this property is an API implementation detail: API client may specify it to get a "dry run" without the actual action. See https://google.aip.dev/163#guidance

### Expected Behavior

The parameter should not be projected to SDKs. Instead, the provider should set it to false (or leave unset) automatically.

### Steps to reproduce

```
new google.eventarc.v1.Trigger("primary", {
    location: "europe-west1",
    eventFilters: [{
        attribute: "type",
        value: "google.cloud.pubsub.topic.v1.messagePublished",
    }],
    labels: {
        foo: "bar2",
    },
    destination: {
        cloudRun: {
            service: "service-1ea8388",
            region: "us-central1",
        },
    },
    triggerId: "mytrigger",
    validateOnly: false,  // This is wrong
});
```

### Output of `pulumi about`

Google native v0.30.0

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #866, opened 2023-04-25
Remove validateOnly from SDKs and resource URLs

mikhailshilkov:
Previously, `validateOnly` was a property of the `eventarc.Trigger` resource (and some others). However, this property is an API implementation detail: API client may specify it to get a "dry run" without the actual action. See https://google.aip.dev/163#guidance

This PR ignores all query parameters with this specific name while building SDKs and resource URLs. As a result, we will never set it and it will stay `false` by default.

Fix https://github.com/pulumi/pulumi-google-native/issues/865

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+Bug fixes:
+
+- Remove `validateOnly` query parameter from SDK properties
+  [#865](https://github.com/pulumi/pulumi-google-native/issues/865)
+
 ## v0.30.0 (2023-04-14)
 Upstream breaking changes:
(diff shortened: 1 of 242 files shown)


thomas11 (commented):
Do we know if people use this in their programs currently, e.g., to test stuff before deploying? Can we release this as a minor version?

mikhailshilkov:
> Do we know if people use this in their programs currently, e.g., to test stuff before deploying? Can we release this as a minor version?

@thomas11 Probably, somebody has set it at least where it's erroneously marked as required (e.g. `eventarc.Trigger`). However, that resource doesn't work properly anyway (updates and deletes are broken due to `validateOnly`).

I don't think anybody uses it as the API intends: if you run it with `true`, you will get your resource in Pulumi but not in Google Cloud which is very weird and unsupported behavior.

Minor 0.x versions can be breaking in semver, it's not a problem.

danielrbradley (approved):

PR #912, opened 2023-07-19
Regenerate sdk

iwahbe:
I believe that this is necessary to resolve https://github.com/pulumi/pulumi-google-native/issues/907.

I muddled through the errors as best I could, but I don't really know what I'm doing here. I would apply close scrutiny to https://github.com/pulumi/pulumi-google-native/pull/912/commits/4bef8933056d54b4fa58f32a5ada771d5c2a4f00.


provider/pkg/gen/overrides.go
```diff
@@ -82,6 +87,13 @@ var resourceNameByPathOverrides = map[string]string{
 	"essentialcontacts_v1.json:v1/folders/{foldersId}/contacts":             "FolderContact",
 	"essentialcontacts_v1.json:v1/organizations/{organizationsId}/contacts": "OrganizationContact",
 
+	// GKEHub
+	"gkehub_v1alpha.json:v1alpha/projects/{projectsId}/locations/{locationsId}/memberships/{membershipsId}/rbacrolebindings": "",
```
danielrbradley, line 91: Why are these names blank? Does leaving it blank mean the resource gets skipped?

PR #923, opened 2023-08-31
Add codecov to this repo

mjeffryes:
Modifies `make test_provider` to collect coverage profile files and adds a ci action to push them to codecov.

An example of the codecov analysis can be see in the comment below and at https://app.codecov.io/gh/pulumi/pulumi-google-native/commit/e2e535ad17dd7eee6a5509a3495a0a6dad2ca7b3


mjeffryes:
Codecov has apparently updated to their commenting bot to not try to make diffs when there's no previous history, but you can still see the data in their UI

thomas11 (approved):

PR #926, opened 2023-09-07
Roll out Python wheel build and publishing

t0yv0:
Fixes https://github.com/pulumi/pulumi-google-native/issues/897

As part of https://github.com/pulumi/home/issues/2883 we are enabling:

- generating pyproject.toml instead of setup.py
- building wheels alongside source distributions
- making sure both wheels and source distributions are published on pypi

If all goes well this feature-* PR should result in a PyPI version that can be checked to have both sdist and wheel distros published.


iwahbe:
Looks like there was some SDK drift.

t0yv0:
Actually the right thing to do here is probably fixup https://github.com/pulumi/pulumi-google-native/pull/925 and land it first, and then do this on top cleanly. I'll try to do that.

t0yv0:
https://pypi.org/project/pulumi-google-native/0.31.2a1694203469/#files wheels in evidence! It's working.

EronWright (approved):

Issue #401, opened 2022-03-20
Stateful Instance Group policy external/internal ips accepts Mapping[str, str] and not str


Tahvok:
This affects compute instance groups beta and alpha.

There is an input class `StatefulPolicyPreservedStateArgs` and an output class `StatefulPolicyPreservedStateResponse`
They have two optional parameters `external_ips` and `internal_ips`. Currently they wrongly assume that they accept a `Mapping[str, str]`.
However, that is wrong. It should be `Mapping[str, Mapping[str, str]]` (the second value is a map of str ,str).
Or even better, the second value should be a class of its own that accepts an `auto_delete` parameter.
See here on how to set it properly:
 https://cloud.google.com/compute/docs/instance-groups/configuring-stateful-ip-addresses-in-migs#configuring_stateful_ip_addresses_on_mig_creation

From what I understand, both of this files (outputs.py and _inputs.py) are generated somehow - but I couldn't figure out how, so didn't submit a pr with my fix.
I applied the fix manually to make it work, and it works as intended.

guineveresaenger:
Hi @Tahvok - thank you for filing this issue!

I believe (and @danielrbradley can correct me if I'm wrong) that all code gen lives in https://github.com/pulumi/pulumi-google-native/blob/master/provider/pkg/gen/schema.go. We're still finalizing this provider, so I apologize for the lack of information on how to get started contributing. 

I'd love it if you wanted to open a fix for this and we're happy to help.

PR #963, opened 2024-02-14
Handle additionalProperties

praneetloke:
This should fix https://github.com/pulumi/pulumi-google-native/issues/401.

As stated in the linked issue, the type for the stateful policy of a managed instance group in the compute module should be a mapping of string and an object that accepts the value for the auto-delete rule (enum). The possible values for the auto-delete rule are `NEVER` and `ON_PERMANENT_INSTANCE_DELETION`.

I looked into this briefly and found that the discovery doc for `compute.v1` has the right mapping but my guess is that when generating the schema, the generator is dropping the `$ref` on the `additionalProperties` which turns these into a simple mapping instead of a map of maps.

Here's a snippet from the discovery doc on the `master` branch:

```json
"StatefulPolicyPreservedState": {
      "description": "Configuration of preserved resources.",
      "id": "StatefulPolicyPreservedState",
      "properties": {
        "disks": {
          "additionalProperties": {
            "$ref": "StatefulPolicyPreservedStateDiskDevice"
          },
          "description": "Disks created on the instances that will be preserved on instance delete, update, etc. This map is keyed with the device names of the disks.",
          "type": "object"
        },
        "externalIPs": {
          "additionalProperties": {
            "$ref": "StatefulPolicyPreservedStateNetworkIp"
          },
          "description": "External network IPs assigned to the instances that will be preserved on instance delete, update, etc. This map is keyed with the network interface name.",
          "type": "object"
        },
        "internalIPs": {
          "additionalProperties": {
            "$ref": "StatefulPolicyPreservedStateNetworkIp"
          },
          "description": "Internal network IPs assigned to the instances that will be preserved on instance delete, update, etc. This map is keyed with the network interface name.",
          "type": "object"
        }
      },
      "type": "object"
    },
```

And here's the Pulumi schema for `v1`:

```json
"google-native:compute/v1:StatefulPolicy": {
            "properties": {
                "preservedState": {
                    "type": "object",
                    "$ref": "#/types/google-native:compute/v1:StatefulPolicyPreservedState"
                }
            },
            "type": "object"
        },
        "google-native:compute/v1:StatefulPolicyPreservedState": {
            "description": "Configuration of preserved resources.",
            "properties": {
                "disks": {
                    "type": "object",
                    "description": "Disks created on the instances that will be preserved on instance delete, update, etc. This map is keyed with the device names of the disks."
                },
                "externalIPs": {
                    "type": "object",
                    "description": "External network IPs assigned to the instances that will be preserved on instance delete, update, etc. This map is keyed with the network interface name."
                },
                "internalIPs": {
                    "type": "object",
                    "description": "Internal network IPs assigned to the instances that will be preserved on instance delete, update, etc. This map is keyed with the network interface name."
                }
            },
            "type": "object"
        },
```

Note how all properties of `StatefulPolicyPreservedState` are missing the additionalProperties definition. (I guess the code generator is defaulting properties of `type: object` without any other attributes as maps?)

This PR handles `additionalProperties` for both cases where a property is a map with a primitive value or one where the value is a reference to a type.


provider/pkg/gen/schema.go
```diff
@@ -1386,6 +1386,21 @@ func (g *packageGenerator) genTypeSpec(typeName, propName string, prop *discover
...
 			Ref:  referencedTypeName,
 		}, nil
+	case prop.Type == "object" && prop.AdditionalProperties != nil:
+		// The prop is a map with a string key and a simple value
+		// if it doesn't have a ref.
+		if prop.AdditionalProperties.Ref == "" {
+			return &schema.TypeSpec{
+				Type: "object",
+				AdditionalProperties: &schema.TypeSpec{
+					Type: "string",
```
praneetloke, line 1396: This should not be hard-coded to `string`. Ideally, this uses `prop.AdditionalProperties.Type` but that leads to some errors that I didn't look deeply into. But then again, perhaps a `string` is ok here for now?

praneetloke:
Ping.

provider/pkg/gen/schema.go
mjeffryes: it's true that these are usually strings, but I think if you instead recurse on prop.AditionalProperties, you can get the full type.

provider/pkg/gen/schema.go
```diff
@@ -1386,6 +1386,21 @@ func (g *packageGenerator) genTypeSpec(typeName, propName string, prop *discover
...
+				},
+			}, nil
+		}
+
+		// Otherwise, the value in-turn is a complex type.
+		typePropName := fmt.Sprintf(`%s%s`, typeName, strings.Title(propName))
+		return g.genTypeSpec(typePropName, propName, prop.AdditionalProperties, isOutput)
```
mjeffryes, line 1403: I think this ends up just return the type of the value rather than a map from string -> valueType

mjeffryes (commented):
First off: thanks for contributing this PR! And apologies that it's taken us so long to review. 

Conceptually, I think this is about right, but there are a couple nuances I'd like to make sure we get correct. And you'll also need to run a full `make build` to regenerate all the sdks before we can run tests.

FWIW, we recently made a similar change in the generator for the aws-native cloud control provider; you might look at this part for some inspiration/comparison: https://github.com/pulumi/pulumi-aws-native/pull/1342/files#diff-a462bb9d1699adb657844ae5c68deb17b7edfd49a2a39ab28f7f53e7e8a380b7

mikhailshilkov:
@praneetloke Are you still planning to address Matt's suggestions here?

praneetloke:
@mikhailshilkov I am going to try to get to this this week. If not, I'll close it. I really think this should be finished though, to improve the quality of this native provider. As it stands right now, the bug this is trying to fix corrects a lot of inaccuracies in the schema as you can see in the diff.

provider/pkg/gen/overrides.go
```diff
@@ -188,6 +189,40 @@ var resourceNamePropertyOverrides = map[string]string{
...
+		//
+		// The property is incorrectly referring to its parent type
+		// causing a recursion. This recursion is accidental as the
+		// description of the property suggests that it's a map
+		// of the protocol and the address. So we override the
+		// type spec for this property with the correct
+		// AdditionalProperties type spec.
+		//
+		// Discovery doc: https://servicemanagement.googleapis.com/$discovery/rest?version=v1
+		"overridesByRequestProtocol": {
```
praneetloke, line 208: Hopefully, the above comment is self-explanatory. If you have a different approach to fixing this or think that the recursive type is actually valid in this case, please let me know and I can revert this. However, that would mean that a solution for the recursive reference would be needed for this provider's schema (perhaps similar to k8s' ObjectMeta?)

praneetloke: I discovered this issue when I ran `make build`. Here's how the property is defined in the discovery doc:

```json
"BackendRule": {
...
...
    "overridesByRequestProtocol": {
        "additionalProperties": {
          "$ref": "BackendRule"
        },
        "description": "The map between request protocol and the backend address.",
        "type": "object"
     },
...
...
}
```

praneetloke:
@mjeffryes @mikhailshilkov I've updated this PR. I've separated the SDK regen as a separate commit[1]. I think I have addressed the comments. I also uncovered a problem in the discovery doc for `servicemanagement`. Please have a look at that.

[1]=It's a large number of changes. But let's be honest, this is not the largest number of changes I've made in a PR 😄

provider/pkg/gen/schema.go
```diff
@@ -1386,6 +1396,51 @@ func (g *packageGenerator) genTypeSpec(typeName, propName string, prop *discover
...
+			case "any":
+				return &schema.TypeSpec{
+					Type:                 "object",
+					AdditionalProperties: &schema.TypeSpec{Ref: "pulumi.json#/Any"},
+				}, nil
+			case "array":
+				// An array type for AdditionalProperties doesn't make sense
+				// since it's already a map and multiple entries are
+				// allowed by definition. So generate a type
+				// based on the type of the "items" in the array.
```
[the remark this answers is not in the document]
praneetloke: > a `map[string][]T` is a perfectly valid type (eg. you might have a map from team name to employees list).

Yes. But I wasn't referring to that :) My confusion was about `type: array` and what that is referring to. I guess I didn't realize that the `type: array` here is indicating a list of values for each key. If that is the case, then yeah what I have here is definitely wrong. I'll change this.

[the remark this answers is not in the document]
praneetloke: I've fixed this now.

provider/pkg/gen/overrides.go
praneetloke: > The safest thing to do would probably be to omit this property entirely until someone can take the time to explore the behavior of this api

This is done.

provider/pkg/gen/schema.go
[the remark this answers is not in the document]
praneetloke: By the way, the reason why the array type didn't make sense to me in the first place in this context is, if you look at the UI for creating a budget, you can only have a single value for a key: 
image

Having said that I agree that this can't be generalized like how I did at first because there could be a `<string, any[]>` elsewhere in the future that this would break.

mjeffryes:
Thanks for updating this @praneetloke; I think we still need to address https://github.com/pulumi/pulumi-google-native/pull/963#discussion_r1589546457 and add some test cases to wrap this up.

mjeffryes:
Builds passed (finally)!

mjeffryes (approved):

